PR #63, opened 2015-01-01
Fixed compile error

mathewprokos:
Removed use of &str::to_ascii in decode_hex

### DIFF
--- a/src/lib/hex.rs
+++ b/src/lib/hex.rs
@@ -17,9 +17,10 @@ pub fn decode_hex(hex: &str) -> Vec<u8> {
     let mut r: Vec<u8> = Vec::new();
     let mut pos = 0;
     loop {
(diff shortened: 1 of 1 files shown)


stepancheg:
I've merged your PR, thanks.

However, there's another compilation error due to recent rust changes, that's not easy to fix.

mathewprokos:
I see that. looking at it now =/

Pretty new to rust so taking a bit to grok what is going on.

something about using not being able to use self in an object-safe trait
because the type of self could change.

stepancheg:
No, it's not because type of self could not change. It's because static methods could not be called only using concrete type.

I think Rust is wrong in that place. In my opinion, error should arise at place where you are trying to call these functions, not at place you are declaring a variable of trait type.

stepancheg:
Almost done it.

stepancheg:
Finally fixed in fbe150b89638a4b2d1bbc5319a05a7f502bf6cdf.

mathewprokos:
Awesome!

PR #114, opened 2015-05-27
Add @generated to generated .rs files

rajatgoel:
This will suppress generated code from showing up in phabricator diffs


stepancheg:
I slightly modified your patch, regenerated the code, and pushed. Thanks!

PR #115, opened 2015-06-10
Silence warning about non_snake_case.

Blei:


stepancheg:
You did not regenerate bundled .proto files. I did that.

BTW, rust-protobuf probably should transform camelCase to snake_case during codegeneration.

Blei:
Ah, yeah I forgot that, thanks!

Sounds complicated, this could result in annoying naming conflicts. But it
would definitely fit rust's naming scheme more, and there's precedent as go
does the same thing (well, they change to CamelCase).

2015年6月10日(水) 22:31 Stepan Koltsov <notifications@github.com>:

PR #118, opened 2015-06-19
Added example of basic I/O

wackywendell:
When I first came across your library,I quickly caught on as to how to create a `.proto` file, transform it into a rust module, and use its messages as objects; however, I was still a little confused as to how to dump it to a buffer. As I played around and tested it, I realized I could make that an example file, so I did that; this is the result.


davideagen:
This PR looks great. The library needs some examples like this.

Stii:
Fantastic! Any reason it hasn't been merged? Would help a lot.

Warfields:
^

PR #123, opened 2015-07-27
Input Stream "Raw Bytes" Optimizations

jamwt:
Speed up handling of nontrivial "raw_bytes" reads within the input stream (a big need we have at dropbox).

 * Switch to direct reader-to-slice I/O and bypass the buffer for read sizes greater than the buffer
 * Switch to `bytes` crate for managing buffering, bytes remaining, start/end positions, using `copy_nonoverlapping` to slices.. it takes care of a lot of that for us and eliminates some moving pieces. 
 * Add a performance test for 1k byte arrays and 100k byte arrays.

Results:

 * 1k byte arrays are ~4x faster
 * 100k byte arrays are ~10x faster

Ran tests, and have done lots of loaded profiles on internal daemons here at Dropbox.


src/lib/stream.rs
```diff
@@ -278,6 +320,25 @@ impl<'a> CodedInputStream<'a> {
...
+            unsafe { mem::transmute(bytes) }
+        )
+    }
+
+    #[cfg(target_endian = "little")]
+    pub fn read_raw_little_endian64(&mut self) -> ProtobufResult<u64> {
+        let mut bytes = [0u8; 8];
+        try!(self.read_raw_bytes_immediate(&mut bytes));
+        Ok(
+            unsafe { mem::transmute(bytes) }
```
stepancheg: Is it safe to transmute like that? `bytes` array is guaranteed to be aligned properly?

I'd write it like this:

```
let mut r = 0u64;
let mut bytes = unsafe { mem::transmute(&mut r) };
...
r
```

BTW, it could probably be a little faster if data is initialized with `mem::uninitialized()`.

src/lib/stream.rs
```diff
@@ -291,6 +352,7 @@ impl<'a> CodedInputStream<'a> {
         )
     }
 
+    #[cfg(target_endian = "big")]
```
stepancheg: `big` and `little` should be on function, and at the end `.to_le()` function should be called to swap bytes on big-endian and do nothing on little-endian.

src/perftest/perftest_data.rs
```diff
@@ -1810,121 +2085,143 @@ static file_descriptor_proto_data: &'static [u8] = &[
...
-    0x3b, 0x3c, 0x0a, 0x0b, 0x0a, 0x04, 0x04, 0x06, 0x02, 0x04, 0x12, 0x03, 0x23, 0x04, 0x2a, 0x0a,
-    0x0c, 0x0a, 0x05, 0x04, 0x06, 0x02, 0x04, 0x04, 0x12, 0x03, 0x23, 0x04, 0x0c, 0x0a, 0x0c, 0x0a,
-    0x05, 0x04, 0x06, 0x02, 0x04, 0x06, 0x12, 0x03, 0x23, 0x0d, 0x18, 0x0a, 0x0c, 0x0a, 0x05, 0x04,
-    0x06, 0x02, 0x04, 0x01, 0x12, 0x03, 0x23, 0x19, 0x25, 0x0a, 0x0c, 0x0a, 0x05, 0x04, 0x06, 0x02,
-    0x04, 0x03, 0x12, 0x03, 0x23, 0x28, 0x29, 0x0a, 0x0b, 0x0a, 0x04, 0x04, 0x06, 0x02, 0x05, 0x12,
-    0x03, 0x24, 0x04, 0x44, 0x0a, 0x0c, 0x0a, 0x05, 0x04, 0x06, 0x02, 0x05, 0x04, 0x12, 0x03, 0x24,
-    0x04, 0x0c, 0x0a, 0x0c, 0x0a, 0x05, 0x04, 0x06, 0x02, 0x05, 0x06, 0x12, 0x03, 0x24, 0x0d, 0x24,
-    0x0a, 0x0c, 0x0a, 0x05, 0x04, 0x06, 0x02, 0x05, 0x01, 0x12, 0x03, 0x24, 0x25, 0x3f, 0x0a, 0x0c,
-    0x0a, 0x05, 0x04, 0x06, 0x02, 0x05, 0x03, 0x12, 0x03, 0x24, 0x42, 0x43,
+    0x28, 0x09, 0x12, 0x0a, 0x0a, 0x02, 0x73, 0x33, 0x18, 0x03, 0x20, 0x01, 0x28, 0x09, 0x22, 0x17,
```
stepancheg: You should not commit this part. It is changed just because protobuf on your host is of different version than on mine. No .proto-files changed by this commit.

stepancheg:
And, could you please avoid dependency on `bytes` crate? I'm keeping `rust-protobuf` with zero dependencies, it is much easier to maintain it that way (for example, no need to worry about backward-incompatible changes in dependent crates).

Probably parts of `bytes` crate should be copied into `rust-protobuf` repo.

And, if you do the changes, could you please squash commits in PR into one or two?

jamwt:
Sure, I can make these changes--thanks for the review.

About the bytes crate, though--can't that just be addressed by pinning the dependency?  Seems odd and not very idiomatic to work around the package management system..

stepancheg:
@jamwt we can't just pin the dependency, because it would create a problem for `rust-protobuf` users who need newer version of `bytes` crate, explicitly or via another library that depends on `bytes` crate.

stepancheg:
It is a little odd indeed, but at least internals of that crate are not exposed as part of rust-protobuf public API.

src/lib/stream.rs
jamwt: Nah, rustc generates:

```llvm
  %a = alloca [8 x i8]
```

... without an alignment, so it's up to the backend to choose a default.  Which, may or may not be word-aligned, so this is probably not safe on all platforms (though I think it probably happens to be on x64).  I'll change it.

jamwt:
Okay.  This seems a little like the larger debate over whether semver can work at all in the large, but I'll look into extracting some `bytes` code and adding Carl's license somewhere if that's the direction you'd prefer to take.

jamwt:
Latest push responds to feedback:

 * Move endianness handling into separate function (with tested flip functions)
 * Ensure stack alignment of u64 by making primary variable a stack u64, and aliasing the slice on top of it
 * Bundle a version of bytes and remove dependencies
 * Squash all commits into one.

src/lib/stream.rs
```diff
@@ -567,22 +601,11 @@ impl<'a> CodedInputStream<'a> {
         self.read_unknown(wire_type).map(|_| ())
     }
 
-    /// Read raw bytes into supplied vector. Vector must be empty.
```
stepancheg: It should say now something like "... Vector is cleared first".

src/lib/stream.rs
```diff
@@ -205,39 +157,140 @@ impl<'a> CodedInputStream<'a> {
...
+        buf.advance(1);
+        Ok(v)
+    }
+
+    /// Read raw bytes into supplied slice, potentially many buffers long.
+    pub fn read_raw_bytes_into(&mut self, count: u32, target: &mut[u8]) -> ProtobufResult<()> {
+        if count == 0 {
+            return Ok(())
+        }
+        assert!(count as usize <= target.len());
```
stepancheg: `count` parameter is redundant. `target` slice can be easily made of `count` size by caller.

src/lib/stream.rs
```diff
@@ -205,39 +157,140 @@ impl<'a> CodedInputStream<'a> {
...
-            self.buffer_size_after_limit = buffer_end - self.current_limit;
-            self.buffer_size -= self.buffer_size_after_limit;
-        } else {
-            self.buffer_size_after_limit = 0;
+    pub fn read_raw_bytes_immediate(&mut self, buf: &mut[u8]) -> ProtobufResult<()> {
+        let count = buf.len();
+        if count == 0 {
+            return Ok(())
+        }
+        assert!(count <= 4096);
```
stepancheg: `4096` constant is used more than once now. So it needs a name.

src/lib/stream.rs
```diff
@@ -205,39 +157,140 @@ impl<'a> CodedInputStream<'a> {
...
+                w.advance(written);
+                if w.has_remaining() {
+                    return Err(ProtobufError::IoError(io::Error::new(
+                        io::ErrorKind::Other, "unexpected EOF")));
+                }
+            },
+            &mut InputSource::Reader(ref mut reader, ref mut b) => {
+                let written = b.read_slice(w.mut_bytes());
+                w.advance(written);
+                while w.has_remaining() {
```
stepancheg: This part is tricky.

If the buffer is small, there's a chance that expensive `.read()` call (e. g. `read` syscall) will be called will small buffer (e. g. 7 bytes). These situations are rare (because `read_raw_bytes_into` are never called sequentially, it is always preceded by reading length which fills the buffer), but possible.

I suggest to explain it in comments, and possibly (I'm not sure) fill the buffer if `target` is smaller than 4096.

src/lib/endian.rs
```diff
@@ -0,0 +1,58 @@
+#[cfg(target_endian = "little")]
+pub fn to_le_u64(v: u64) -> u64 {
```
stepancheg: These functions exist [in the rust standard library](http://is.gd/WMoDc0), no need to reimplement them.

stepancheg:
I've got another idea.

Probably rust-protobuf should not do buffering itself, and just require that reader is already `BufRead` from rust standard library that seems to do buffering pretty similar to what you've implemented (at least common `BufReader` implementation). It would also avoid double buffering when `Read` is already buffered.

When I completed first version of protobuf, rust didn't have `BufRead` trait.

However it would require throwing away all your work.

BTW, have a look at [topic I've started on rust-users](https://users.rust-lang.org/t/read-is-bufread-or-not/2256).

src/lib/endian.rs
jamwt: Didn't know, thanks.

src/lib/stream.rs
jamwt: So, the idea here is that if your reader involves expensive syscalls, it's up to you to pass in a `BufRead`.  Taking `Read`, and leaving those kinds of decisions up to the caller, seems like the right approach rather than dictating `BufRead`.

xitep:
I haven't followed all the details, but as far as I can tell, having users to supply a "buffered reader" should be the standard approach. I'd not be very happy if rust-protobuf did (unnecessarily) internal buffering and copying when I supply it an already in-memory byte slice to parse.

jamwt:
Ah, I've been assuming the wrong thing here.  I assumed that BufRead is a reading adapter that just implements `Read` like golang's equivalent (that you could use interchangeably for `Read`) rather than a completely separate trait.  In that case, I agree this whole thing should probably just take a `BufRead`.

jamwt:
@xitep in-memory slices are not double buffered.

src/lib/endian.rs
jamwt: Had no idea.  Thanks.  I googled, but didn't turn anything up.

jamwt:
@stepancheg Actually, `BufReader` does implement `Read`.  So if you want to avoid expensive syscalls, you can do that with a `BufReader` and the interface can still take `Read`.

I do agree about avoiding the double buffering, though; I just preserved that 4k array b/c I assumed that was based on some other tuning you had done or some goal you had in mind.  A simpler approach would to either always use the slice or to always use the reader and to allow the caller to do the smart thing wrt using `BufReader` (as opposed to requiring `BufRead` trait) where the application or reader type demands it.

xitep:
@jamwt ah, i see. thanks for the explanation.

jamwt:
@stepancheg New proposal:

I rework this to always have the consumption be regular reads on a `Read`.  The `Read` is one of two things:

 1. an io::Cursor(&[u8]) if a slice was given
 2. Whatever `Read` was passed in by the caller.  If they need to avoid syscalls, it is on them to use a `BufReader` or whatever other mechanism they desire.

And then we get rid of the internal 4k buffer, and drop the `bytes` dependency.

jamwt:
Okay, so dropping the internal buffering is somewhat complicated by the `eof()` checks throughout the code.  The library makes pretty liberal use of checking for a zero-length read on the socket, and dumping the body into the buffer if on the non-eof condition... they making that buffer seamlessly available as part of the input stream in subsequent read calls.

jamwt:
Maybe that's the argument for requiring `BufRead`?  We can use the `fill_buf()` method to check eof, but a subsequent `read()` will still do the right thing?

jamwt:
Okay, another change!

Threw out a bunch of code.  Reworked everything around `BufRead`.

Commit message copypasta:

-----------

Switch over to using BufRead for buffering instead of maintaining our own internal buffers.

Input sources now have three modes:

 * A `BufRead` is provided directly, and we use it.
 * A slice is provided, and we wrap it in `io::Cursor`, which will
   provide `BufRead`, using the wrapped slice itself for the "filled
   buffer" so we can check for eof.
 * A normal `Read` is provided, in which case we wrap it in a
   `BufReader` with a 4k internal buffer.

This removes a decent amount of code and speeds up handling of raw byte
strings and byte fields quite a bit:

 * ~4x faster for 1k byte ararys
 * ~9x faster for 100k byte arrays

Tests added to support this; the change should be fully
backward-compatible since the default `new()` constructor to
CodedInputStream remains `Read` based.

src/lib/stream.rs
```diff
@@ -567,22 +542,12 @@ impl<'a> CodedInputStream<'a> {
         self.read_unknown(wire_type).map(|_| ())
     }
 
-    /// Read raw bytes into supplied vector. Vector must be empty.
+    /// Read raw bytes into the supplied vector.  The vector will be resized as needed and
+    /// overwritten.
     pub fn read_raw_bytes_into(&mut self, count: u32, target: &mut Vec<u8>) -> ProtobufResult<()> {
-        assert!(target.is_empty());
         target.reserve(count as usize);
```
stepancheg: This doesn't work properly now if `target` is not empty, because `reserve` reserves _additional_ bytes. You need to either return the assertion or `.set_len(0)` first.

stepancheg:
> I do agree about avoiding the double buffering, though; I just preserved that 4k array b/c I assumed that was based on some other tuning you had done or some goal you had in mind.

Actually, yes, internal buffer is somewhat faster, because it avoids unnecessary non-inlinable virtual calls (in case of `InputSource::Read` variant). I don't know if is it a big deal or not.

stepancheg:
Fixed `set_len` and pushed 19fff8f934b6f00443a123d7ad3537dd1c8bf9fa.

Thanks a lot! If you have ideas what else needs optimization, please share!

jamwt:
Thank you!

PR #169, opened 2016-07-05
Fix incorrect varint length for negatives numbers

dvvc:
Use a signed number absolute value when calculating its varint size, instead of casting it to integer which causes `std::T::MAX + 1` to be added to it when it's negative.


dvvc:
This patch is no good, since it breaks int behavior (it should only change size calculation for sint values). Looking for a fix.

PR #195, opened 2017-01-14
Problem: not using lib.rs rust convention causes a level violation

sjmackenzie:
or state leakage in that build systems such as nixcrates don't
know the library entry point convention. It would require
users of such systems to explicitly name the entry point for every crate.

Solution: rename /lib/protobuf.rs -> /lib/lib.rs


sjmackenzie:
harrah!

PR #463, opened 2019-12-13
Edition 2018: fix idioms

NickeZ:
I'm having some problems with protobuf and cbindgen. `syn` cannot parse the crate because it uses an old way of writing trait functions.

See https://github.com/dtolnay/syn/issues/733

As a first step I've ran `cargo fix --edition-idioms`.


dtolnay:
I don't think there is agreement that the changes made by --edition-idioms are idiomatic unfortunately. In all of the changes in this PR I would prefer to keep the old code even after the edition is updated to 2018.

PR #492, opened 2020-05-28
Use full name of Box for generated into_any fn.

### LINKED ISSUE #491: Enhancement: self defined `Box` type is conflict with `std::boxed::Box`
If protocol buffer files defined a type `Box`, the generated source file can't be compile for different types:

```
9188 |     fn into_any(self: Box<Self>) -> ::std::boxed::Box<dyn (::std::any::Any)> {
     |                           ^^^^ unexpected type argument

error[E0107]: wrong number of type arguments: expected 0, found 1
```

Ver: 2.14

liubin:
This will avoid conflict if protocol file also defined a message named Box.

Signed-off-by: bin liu <bin@hyper.sh>

This will fix #491 at v2.14

### DIFF
--- a/protobuf-codegen/src/message.rs
+++ b/protobuf-codegen/src/message.rs
@@ -369,7 +369,7 @@ impl<'a> MessageGen<'a> {
                 w.write_line("self as &mut dyn (::std::any::Any)");
             });
             w.def_fn(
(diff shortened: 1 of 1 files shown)


stepancheg:
Thanks!

stepancheg:
Published version 2.15.1.

PR #537, opened 2021-01-06
Upgrade to bytes v1.0.0

giarc3:


giarc3:
I wasn't compiling with the correct feature flags, I'll fix and re-open

PR #550, opened 2021-03-05
Check "Always Output Default Values" for Repeated Fields in JSON Output

wtachau:


wtachau:
Hi @stepancheg , let us know if there are any changes you'd like us to make

wtachau:
@stepancheg anything you'd like me to update here? thanks!

PR #657, opened 2022-10-26
Add support for wrapping message in Option 

overvenus:
`MessageField` has a noticeable overhead where there are lots of small nested messages being created and dropped frequently, because it wraps messages in `Box`.

This PR add an option `option_for_message` to rustproto that allows to wrap a message in an `Option` in generated code.

Also adding a simple benchmark shows the improvement. 

```
# Results in Intel(R) Xeon(R) CPU E5-2630 v4 @ 2.20GHz
test clone_and_drop_message_field             ... bench:  148 ns/iter (+/- 24)
test clone_and_drop_message_field_with_option ... bench:   36 ns/iter (+/- 5)
```

I would appreciate feedback!


MzChynaBlackInc (approved):
f10fe57a3a1b4cedafbae84b408e4177d343fa3f

PR #660, opened 2022-12-05
MessageDyn: Do not omit zero values on fields containing oneof

mingyuchi:
Omitting zeros on singular fields only make sense if it's not in oneof
For example, omitting zero on the following oneof will make it ambiguous
```proto
oneof int_or_string {
  int32 int = 1;
  string str = 1;
}
```

if either int or str is zero, both of the fields will be missing.
Now it's not possible to know which field was originally filled


SeongMinPark-devsisters:
@stepancheg Hello, what do you think of this pr?

PR #661, opened 2022-12-21
Implement conformance test binary

Ralith:
Draft implementation of the conformance test protocol defined at https://github.com/protocolbuffers/protobuf/tree/main/conformance. Not yet tested.


Ralith:
I think this is working now. It's found a bunch of failures when ran against the test suite from 3.21. Looks like a lot of unexpected successful parses (particularly around EoF handling), but also some panics and many cases of data failing to pass through. Haven't fully eliminated the possibility that the test is doing something wrong.
[test-results.txt](https://github.com/stepancheg/rust-protobuf/files/10274964/test-results.txt)

PR #696, opened 2023-12-21
Codegen: option to use BTreeMap as map representation

akhramov:
As of now, protobuf maps are represented with
[std::collections::HashMap](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html), which serves as a robust default. In rarer instances, opting for a different implementation, such as BTreeMap, might be desirable, e.g. for deterministic serialization (#496).

This change

* adds `btreemaps` codegen option to generate [std::collections::BTreeMap](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html) for maps representation.


test-crates/protobuf-codegen-btreemap-test/Cargo.toml
akhramov, line 1: This is basically a copy of [test-crates/protobuf-codegen-protoc-test](https://github.com/stepancheg/rust-protobuf/tree/a6f27be0dbdcb5e88182a3a289a93179c994fa72/test-crates/protobuf-codegen-protoc-test)

test-crates/protobuf-codegen-btreemap-test/build.rs
```diff
@@ -0,0 +1,90 @@
...
+        |GenInDirArgs {
+             out_dir,
+             input,
+             customize,
+         }| {
+            codegen()
+                .out_dir(out_dir)
+                .inputs(input)
+                .includes(&["../../proto", include_dir])
+                .customize(customize.btreemaps(true))
```
akhramov, line 33: Unlike protobuf-codegen-protoc-test, we generate btreemaps

test-crates/protobuf-codegen-btreemap-test/src/common/v2/test_reflect_clear.rs
```diff
@@ -0,0 +1,71 @@
+use std::collections::BTreeMap;
```
akhramov, line 1: The last difference w. protobuf-codegen-protoc-test

PR #713, opened 2024-02-18
Add new end-to-end example

jgarzik:
Five years ago, I created a fork-and-go working protobufs example: https://github.com/jgarzik/rust-protobuf-example  Periodically, users will request that this example be merged into this upstream repo.

This PR adds this full end-to-end working example.

The goal in this example is to demonstrate the common case of encoding and decoding a set of mock API requests.

Happy to revise this as requested by the rust-protobuf maintainers.


jgarzik:
Two linter-related checks appear to fail for reasons unrelated to my PR?

Mega-linter needs an upgrade?

Issue #742, opened 2024-09-29
compile error with `required group` proto2 message


kristopherbullinger:
```proto
// protos/group.proto
syntax = "proto2";
message Foo {
  required group Data = 201 { // <------- `required group`
    required int32 group_int32 = 202;
    required uint32 group_uint32 = 203;
  }
}
```

```rs
// build.rs
fn main() {
    println!("cargo:rerun-if-changed=protos/group.proto");
    protobuf_codegen::Codegen::new()
        .pure()
        .includes(&["protos"])
        .input("protos/oneof.proto")
        .cargo_out_dir("protos")
        .run_from_script();
}
```

```rs
// src/main.rs
mod protos {
    include!(concat!(env!("OUT_DIR"), "/protos/mod.rs"));
}
use protos::group;
fn main() {}
```

Produces compile error:
```
error[E0609]: no field `data` on type `&Foo`
  --> /protobuf-build-test/target/debug/build/oneof-7aa1b7e9780d5143/out/protos/group.rs:63:17
   |
63 |         if self.data.is_none() {
   |                 ^^^^ unknown field
   |
   = note: available field is: `special_fields`
```

It seems that a group is merely rendered as a comment in the generated output struct:
```rs
// @@protoc_insertion_point(message:Foo)
#[derive(PartialEq,Clone,Default,Debug)]
pub struct Foo {
    // message fields
    // data: <group>    <------------------------------
    // special fields
    // @@protoc_insertion_point(special_field:Foo.special_fields)
    pub special_fields: ::protobuf::SpecialFields,
}
```
Perhaps it should be rendered as a `data: Option<foo::Data>,` ?

PR #743, opened 2024-10-02
add support for proto2 groups

kristopherbullinger:
Hello. Thanks for your work on this repo. I'm interested in maximizing support for protobufs in rust.

In the interest of increasing protobuf conformance score, this pull request adds support for the deprecated proto2 group syntax.

Groups are generated similarly to messages; a rust struct is created to match the contents of the group.

Groups use a delimited merge operation when reading from a `CodedInputStream`. To avoid making a change to the public `Message` trait, an inherent method is generated in each group- and message-derived struct called `merge_delimited`, which takes an `end_tag`, and merges into the struct until that end tag is found. The Message impl for a struct leverages the inherent `merge_delimited` method on its message field for groups. Example:
```proto
syntax = "proto2";
message Foo {
  repeated group Data = 1 {
    optional int32 group_int32 = 2;
    required uint32 group_uint32 = 3;
  }
}
```
```rs
    // <Foo as Message>::merge_from
    pub fn merge_from(&mut self, is: &mut ::protobuf::CodedInputStream<'_>) -> ::protobuf::Result<()> { 
        while let Some(tag) = is.read_raw_tag_or_eof()? {
            match tag {
                11 => {
                    let end_tag = ::protobuf::rt::set_wire_type_to_end_group(tag);
                    let mut data = foo::Data::default();
                    data.merge_delimited(is, end_tag)?; // <-- `foo::Data::merge_delimited` inherent method
                    self.data.push(data);
                },
```

When an unknown type is encountered, if the tag is of type StartGroup, a `CodedOutputStream` is created on top of a `Vec<u8>`, and all values are written to it until the corresponding end tag is found. The `CodedOutputStream` is then flushed, and the raw bytes returned for storage.

The computed serialized size of a group is the size of the message plus twice the size of the tag.

Generated code compiles with `repeated`, `required`, and `optional` groups. 

This pull request addresses and provides a fix for [Issue#742](https://github.com/stepancheg/rust-protobuf/issues/742).

This feature addition covers a large surface area and I am happy to incorporate any feedback. Please let me know if I missed anything or if my implementation does not suit the coding style of this project.

### DIFF
--- a/protobuf-codegen/src/gen/field/accessor.rs
+++ b/protobuf-codegen/src/gen/field/accessor.rs
@@ -155,9 +155,11 @@ impl FieldGen<'_> {
                 type_params: vec!["_".to_owned()],
                 callback_params: self.make_accessor_fns_lambda(),
             },
-            FieldElem::Group => {
-                unreachable!("no accessor for group field");
-            }
+            FieldElem::Group(m) => AccessorFn {
+                name: "make_message_field_accessor".to_owned(),
+                type_params: vec![format!("{}", m.rust_name_relative(&self.file_and_mod()))],
+                callback_params: self.make_accessor_fns_lambda(),
+            },
         }
     }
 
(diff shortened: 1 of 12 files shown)


kristopherbullinger:
Lots of tests are failing in the `test-crates` which I didn't use at first. I'll work on getting them to pass.

kristopherbullinger:
All of the tests are passing in each of the test crates. I think the remaining CI failures have something to do with the `protobuf/regenerate.sh` commands. Maybe you can help me get this over the finish line?

PR #762, opened 2025-04-23
Add an option to print 64-bit integers without quotes

svanharmelen:
I appreciate the fact that this is done because it is described as such in the [ProtoJSON](https://protobuf.dev/programming-guides/json/) specification, but it prevents using the generated JSON in (most) other systems as it's not following [RFC8259](https://www.rfc-editor.org/rfc/rfc8259#page-7) which does not specify nor require 64-bit integers to be quoted.

By adding this option people still get the `ProtoJSON` output by default, but have the option to opt-out from the quoted 64-bit integers in order for the output to be more inline with `RFC8259`.

Hope this is something you are willing to merge as it will allow us to use the `protobuf-json-mapping` crate for our use cases. Thanks!


svanharmelen:
I'll have a look at the failed tests...

svanharmelen:
Any updates or feedback on this one?

